PR #102, opened 2017-06-29
Switch to NodePort in example

cdrage:
Simplest of simple should use NodePort


kadel:
nodePort is not really good practice,  I don't think we should show it in our examples

Issue #105, opened 2017-06-29
make test fails


cdrage:
Despite us not having any tests yet, make test / go fmt fails:

```
▶ make test
go get github.com/mattn/goveralls
go get github.com/modocache/gover
go get github.com/Masterminds/glide
go get github.com/sgotti/glide-vc
go get github.com/golang/lint/golint
./scripts/check-vendor.sh
Checking for nested vendor dirs
OK
Checking if vendor was cleaned using glide-vc.
OK
./scripts/check-gofmt.sh
gofmt ERROR - These files are not formated by gofmt:
./cmd/deploy.go
./cmd/generate.go
./pkg/encoding/fix.go
./pkg/transform/kubernetes/kubernetes.go
Makefile:44: recipe for target 'gofmt' failed
make: *** [gofmt] Error 1
```

PR #107, opened 2017-06-30
run unit test and other basic test in travis-ci

kadel:
I know that we will use fabric8 in the future, but until then we need to have some besics check for PRs to prevent some basic mistakes like #105.

This runs some basic test on vendor directory, checks if files are formatted according to gofmt, and runs unit tests.

Currently unit tests with go 1.6 are failing with:
```
pkg/transform/kubernetes/kubernetes_test.go:44: t.Run undefined (type *testing.T has no field or method Run)
```

^ this uses something that is only  in go1.7 and up

I've marked go1.6 as allow failures.
We should consider dropping  support for go1.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+dist: trusty
+
+language: go
+
+# make this also work for forks
+go_import_path: github.com/kedgeproject/kedge
+
+matrix:
+  include:
+    - go: 1.6
+    - go: 1.7
+    - go: 1.8
+  allow_failures:
+    - go: 1.6
+
+install:
+  - true
+
+script:
+  - make test
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ validate: gofmt vet lint
 vet:
 	go vet $(PKGS)
 
+# golint errors are only recommendations
 .PHONY: lint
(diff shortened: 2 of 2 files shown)


kadel:
There is also a lot of golint errors, but those are only suggestions as described in golint readme https://github.com/golang/lint

It would be nice to clean some of them in the future, but it's not something that is that important.

Makefile
```diff
@@ -35,6 +35,7 @@ validate: gofmt vet lint
 vet:
 	go vet $(PKGS)
 
+# golint errors are only recomendations
```
cdrage: recommendations

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+  allow_failures:
+    - go: 1.6
+
+install:
+  - true
+
+script:
+  - make test
+
+
```
cdrage: extra newlines

Makefile
kadel: fixed

.travis.yml
kadel: fixed

cdrage:
CI passes! Merging!

PR #142, opened 2017-07-11
Rename commands from deploy/undeploy to create/delete

cdrage:
Changes the commands to create/delete rather than deploy/undeploy.
Similar to the concepts of `kubectl`.


cdrage:
Hey @pradeepto I've gone ahead and added the documentation changes to this PR.

PR #154, opened 2017-07-13
Upload binaries to Bintray

kadel:
Kedge binary is automatically uploaded to Bintray CDN.

Build and upload happens every time there is a change in master branch.

There is no version history right now. Only the latest version is available.

I'll add a smarter version with history and with posting comments to PRs with link to binary in another PR.

Link to download latest binary is https://dl.bintray.com/kedgeproject/kedge/latest/kedge

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Kedge - Concise Application Definition for Kubernetes
 
-[![Build Status Widget]][Build Status] [![GoDoc Widget]][GoDoc] [![Slack Widget]][Slack]
+[![Build Status Widget]][Build Status] [![GoDoc Widget]][GoDoc] [![Slack Widget]][Slack] 
+
 
 
 ## What is Kedge?
@@ -24,8 +25,12 @@ Check out our [roadmap](ROADMAP.md) as we push towards a __0.1.0__ release.
 ## Using Kedge
 
 ### Installation
+You can download latest master build for all major operating systems and architectures.
+- [Linux (amd64)][Bintray Latest Linux]
+- [macOS (darwin)][Bintray Latest macOS]
+- [Windows (amd64)][Bintray Latest Windows]
 
-The _best_ way to try Kedge is to download the most up-to-date binary from the master GitHub branch:
+Or you can build it yourself from the master GitHub branch:
 
 ```sh
 go get github.com/kedgeproject/kedge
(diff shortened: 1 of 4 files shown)


cdrage:
@kadel Since we're building the binaries, we'd have to build for Linux, macOS and Windows.

The link to the "latest" binary has no indication what platform the latest binary is being built for. We'll have to build for all three :+1:

kadel:
> @kadel Since we're building the binaries, we'd have to build for Linux, macOS and Windows.
> 
> The link to the "latest" binary has no indication what platform the latest binary is being built for. We'll have to build for all three

Ahhh, You are right, I completely forgot about other OSes :blush:

cdrage:
@kadel Could you let me know what CNAME I should set too?

kadel:
> @kadel Could you let me know what CNAME I should set too?

Why CNAME?
We can use direct links to files no? Like this: https://dl.bintray.com/kedgeproject/kedge/latest/kedge

cdrage:
@kadel because it may look nicer with: https://dl.kedgeproject.org/latest/kedge-linux-amd64 or something along that line.

kadel:
I wouldn't worry about that right now ;-)

README.md
```diff
@@ -1,6 +1,8 @@
 # Kedge - Concise Application Definition for Kubernetes
 
-[![Build Status Widget]][Build Status] [![GoDoc Widget]][GoDoc] [![Slack Widget]][Slack]
+[![Build Status Widget]][Build Status] [![GoDoc Widget]][GoDoc] [![Slack Widget]][Slack] 
+
+[![Download Widget Linux]][Bintray Latest Linux] [![Download Widget macOS]][Bintray Latest macOS] [![Download Widget Windows]][Bintray Latest Windows]
```
cdrage: So it seems a bit odd that the widgets at the top have no indication if they're from master or not, or whether it's a tagged release. Maybe a small comment saying

__Latest master binaries:__

What's your opinion?

kadel: I'm even considering removing widgets. I've put them only to see how it looks.

kadel:
> Links ain't working right now (see: https://dl.bintray.com/kedgeproject/kedge/latest/kedge-darwin-amd64) but I assume that's because it needs to be built / triggered first.

Links should work now. I've deleted files, so it was unavailable until job finesed.

> Stupid question, but I'm assuming that the deploy key also means that it's ONLY triggered when a PR is merged?

Yes, it means if there is a change in master branch, either when PR is merged, or if someone pushes something directly to master (but that should never happen) ;-)

cdrage:
LGTM! :+1:

PR #177, opened 2017-07-24
add files passed check to apply.go

containscafeine:
Before this commit, "kedge apply" did not check if files were
being passed to it or not. This commit adds that behavior and now,
"kedge apply" fails when no fails are passed to it using the
"-f/--files" flag.


containscafeine:
@surajssd fixed

PR #194, opened 2017-08-01
Added coverage badge

surajnarwade:
Added coverage badge as well as coverage test in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,10 @@ test-unit:
 .PHONY: test
 test: test-dep check-vendor validate test-unit
 
+# Tests that are run on travs-ci
+.PHONY: travis-tests
+travis-tests: test-dep check-vendor validate test-unit-cover
+
 # Install all the required test-dependencies before executing tests (only valid when running `make test`)
 .PHONY: test-dep
 test-dep:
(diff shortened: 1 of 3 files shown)


surajnarwade:
@kadel , I need to turn on coverage on coveralls, right ?
bit confused here

Makefile
```diff
@@ -56,7 +56,7 @@ test-unit:
 
 # Run all tests
 .PHONY: test
-test: test-dep check-vendor validate test-unit
+test: test-dep check-vendor validate test-unit test-unit-cover
```
kadel: This runs tests twice.

kadel:
It might be better to create new Makefile rule called `travis-tests`

that will look like this:

```Makefile
# Tests that are run on travs-ci
.PHONY: travis-tests
travis-tests: test-dep check-vendor validate test-unit-cover
```

This rule will be used in `.travis.yaml` instead of `make test`.

We can leave `make test` as it was, so it can be easily run by Kedge developers on their local machine.

kadel:
I've disabled coveralls comments in the coveralls settings. It's more annoying than useful :-)

PR #226, opened 2017-08-17
move files from package encoding to package spec

containscafeine:
This commit is purely a refactor. Some files from pkg/encoding
have been moved to pkg/spec. This is the first part of a bigger
effort to consolidate all the controller specific functions and
methods under one package, i.e. package spec for now, so that we
can start defining new controllers on a codebase that's more
suited for extensibility.

### DIFF
--- a/pkg/spec/fix.go
+++ b/pkg/spec/fix.go
@@ -14,18 +14,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package encoding
+package spec
 
 import (
 	"fmt"
 	"strconv"
 
-	"github.com/kedgeproject/kedge/pkg/spec"
-
 	"github.com/pkg/errors"
 )
 
-func fixApp(app *spec.App) error {
+func FixApp(app *App) error {
 
 	// fix app.Services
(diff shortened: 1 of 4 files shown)


kadel:
what is really strange is that coverage decreased, it would be nice to know why

pkg/spec/fix.go
```diff
@@ -14,18 +14,16 @@ See the License for the specific language governing permissions and
...
 	"fmt"
 	"strconv"
 
-	"github.com/kedgeproject/kedge/pkg/spec"
-
 	"github.com/pkg/errors"
 )
 
-func fixApp(app *spec.App) error {
+func FixApp(app *App) error {
```
kadel: why not `func (app *App)FixApp() error`?

containscafeine: Yep, future PRs, once the controller interface concept is in.

containscafeine:
> what is really strange is that coverage decreased, it would be nice to know why

@kadel I think maybe the number of lines and number of packages decreased in the project but the tests did not increase. Not sure :(

containscafeine:
Thanks for reviewing @kadel :)

PR #232, opened 2017-08-21
Fixed port issue in `kedge init`

surajnarwade:
Earlier `kedge init` used to allow non int values for port, Now it will allow only int values.


pradeepto:
@surajnarwade Please add `kedge init` bits to the readme (this different from the asciicinema). There is a "trying me out" section (https://github.com/kedgeproject/kedge#trying-it-out). We should have `kedge init` mentioned there.

PR #255, opened 2017-09-12
move validateVolumeClaims() to resources.go

containscafeine:
This commit moves the incorrectly places validateVolumeClaims()
function from pkg/spec/spec.go to pkg/spec/resources.go.

pkg/spec/spec.go should mainly contain the Kedge spec internal go
definitions, not other functions. This commit fixes that.

### DIFF
--- a/pkg/spec/resources.go
+++ b/pkg/spec/resources.go
@@ -363,3 +363,19 @@ func (app *ControllerFields) CreateK8sObjects() ([]runtime.Object, []string, err
 
 	return objects, app.ExtraResources, nil
 }
+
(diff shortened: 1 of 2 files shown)


containscafeine:
ping @cdrage, any idea why Semaphore is failing? :(

cdrage:
@containscafeine 

Basically, if we use docker implementation of minikube = dns fails. If we use virtualbox of minikube = no vm-in-vm functionality enabled in Semaphore.. :-1: 

Still investigating how to fix this.

containscafeine:
ack

PR #263, opened 2017-09-13
Add timeout param to make test-e2e

cdrage:
Adds a timeout param that can be specified via:

```
TIMEOUT=10m make test-e2e
```


cdrage:
@surajssd added docs.

PR #288, opened 2017-09-21
[WIP] Add DeploymentConfig support

cdrage:
Adds DeploymentConfig support to Kedge by being able to specify a
controller like so:

```yaml
controller: deploymentconfig
```

Unfortunatley due to conflicting vendoring, a "hack" of openshift and
kubernetes types is added in order to prevent the full importation of
kubernetes/kubernetes which uses conflicting vendored files between
client-go, apimachinery and the main kubernetes/kubernetes directory.

Within the `/hack` directory is code copied and pasted from both
Kubernetes and OpenShift in order for us to use only what we need (the
structs).


cdrage:
FINALLY GOT VENDORING TO WORK :100: 

Some todo's as well as notes:

TODO:
 - [ ] Numero uno: DeploymentConfig spits out a blank PodTemplateSpec struct, more support will be added
 - [ ] Tests need to be added (unit, integration)

NOTES:
 - Have a look at `glide.yaml`. I've had to push a fork of apimachinery with a minor edit in order to get this to work
 - A lot of hacking around vendoring, but in the end, I'm able to add OpenShift support without having to add both openshift/origin and kubernetes/kubernetes to Kedge, looking at a smaller binary than what it would of originally have been with both included.

cdrage:
@kadel Do you have any idea what's happening with `go test -i -race -cover ./cmd/... ./pkg/... ./tests/... .` in the Travis tests (failure), I forgot, but is there a way to update the dependencies / tests?

kadel:
How is this going to work when we need update kubernetes or openshift ?

cdrage:
@kadel Since we only use the structs, I don't believe we'll need to update Kubernetes or OpenShift much.

It will only be when `go-client` updates we may run into a minor issue, but it'll be as simple of updating and then patching `cdrage/apimachinery` fork.

Were you able to look at the error shown within Travis?

cdrage:
Ah, looks like I found the failure, it was because I forgot to add `glide-vc --only-code --no-tests --use-lock-file`. Specifically, the `--use-lock-file` part.

I shall now work on adding integration / unit / cmd tests.

kadel:
> @kadel Since we only use the structs, I don't believe we'll need to update Kubernetes or OpenShift much.

We use a bit more than just structs. For example Schema (https://github.com/kedgeproject/kedge/blob/15a202460f3b79e6f8dde14478d609a3a3b2222b/pkg/spec/util.go#L60) If I remember correctly it Schema contains what objects are where (Deployments are in `extensions/v1beta1` before 1.6 and in `apps/v1beta1` after that) 
We should make sure that we can easily update dependencies like kubernetes/client-go and the same for OpenShift deps

cdrage:
@kadel Scheme we only use in order to get the correct controller for Kubernetes.

For example, we explicitly implied it here: 
```go
    TypeMeta: metav1.TypeMeta{                                                                                                                             
      Kind:       "DeploymentConfig",                                                                                                                      
      APIVersion: "v1",                                                                                                                                    
    },
```

However, in the refactor that @containscafeine he removed the TypeMeta portion in order to correctly grab an APIVersion + Kind from Kubernetes.

For OpenShift, we are only using Structs, but for Kubernetes, we're still using Scheme, this can be removed however, but at the moment, upcoming changes upstream to apimachinery wouldn't affect this "hack" in order to retrieve the structs.

kadel:
Generated output doesn't look right :-(

for definition like this
```yaml
name: foobar
controller: deploymentconfig
replicas: 2
strategy:
  type: recreate
containers:
```

output is wrong

```yaml
---
apiVersion: v1
kind: DeploymentConfig
metadata:
  creationTimestamp: null
  labels:
    app: foobar
  name: foobar
spec:
  replicas: 0
  strategy:
    resources:
      Limits: null
      Requests: null
  test: false
  triggers: null
status: {}
```


replicas are always 0 and image is not set and strategy is also not set. It looks like nothing from PodSpec or DeploymentConfigSpec  is propagated


for kedge file like above I would expect otput simillar to this:

```yaml
apiVersion: v1
kind: DeploymentConfig
metadata:
  name: foobar
  labels:
    app: foobar
spec:
  replicas: 2
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app: foobar
      name: foobar
    spec:
      containers:
        - name: foobar
          image: quay.io/tomkral/sleeper
```

pkg/spec/types.go
```diff
@@ -188,6 +194,11 @@ type JobSpecMod struct {
...
 
-type Controller struct {
-	Controller string `json:"controller,omitempty"`
+// TODOOOOOO!!
+// Ochestrator: OpenShift
+// DeploymentConfigSpecMod is Kedge's extension of OpenShift DeploymentConfig in order to define and allow
+// a complete kedge app based on OpenShift
+type DeploymentConfigSpecMod struct {
+	ControllerFields       `json:",inline"`
+	os_v1.DeploymentConfig `json:",inline"`
```
kadel: Should this be `os_v1.DeploymentConfigSpec`?

kadel:
There is really no way to use `openshift/client-go` with `kubernetes/client-go` right now?

cdrage:
@kadel Fixed output is on the TODO list, it's all null since I was just trying to get this working. See my previous comments.

Regarding the libraries:

I tried multiple ways with multiple different commits to see if we were able to get both `apimachinery`, `openshift/origin` and `kubernetes/kubernetes` (specifically `/api`) to all work together. All ended with compiling errors.

Feel free to try, but after two days of trying to work-around it, I used this "hack" instead, but in the end, it resulted in a much smaller binary.

pkg/spec/types.go
cdrage: Good catch, you're right, similar to DeploymentSpec, it should be DeploymentConfigSpec for Openshift. https://github.com/kubernetes/kubernetes/blob/1a44e26670c9ce6b85304ff2603c462cb3a42bca/staging/src/k8s.io/api/apps/v1beta1/types.go#L284

cdrage:
Closing this PR in favour of a new PR: https://github.com/kedgeproject/kedge/pull/315

PR #310, opened 2017-10-04
auto populate ingress name

containscafeine:


containscafeine:
Yikes, I had to add a commit message, nvm now :P

PR #356, opened 2017-10-13
remove Labels from ControllerFields

containscafeine:
This commit removes Labels field from ControllerFields struct
in pkg/types.go.

We do not require Labels because ControllerFields already has
(meta)v1.ObjectMeta struct embedded into it which contains the
Labels field.


kadel:
it looks like this is already done in https://github.com/kedgeproject/kedge/pull/355

containscafeine:
Closing since redundant, already fixed in #355

PR #366, opened 2017-10-17
add support for OpenShift Routes

### LINKED ISSUE #238: Support OpenShift Routes

containscafeine:
This commit adds support for defining OpenShift routes in a Kedge
file.

Routes can be defined like the following at the root level of a
Kedge file

```yaml
routes:
- name: route1
  host: route1-myproject.192.168.42.69.nip.io
  port:
    targetPort: 8080
  to:
    kind: Service
    name: httpd
```

This is implemented by essentially merging RouteSpec and ObjectMeta
and adding an array of the resulting struct at the root level under
the field "routes"

Adds docs, tests.

Fixes #238

### DIFF
--- a/docs/examples/routes/httpd.yml
+++ b/docs/examples/routes/httpd.yml
@@ -1,0 +1,12 @@
+name: httpd
+containers:
+- image: bitnami/nginx
+services:
+- name: httpd
+  type: NodePort
+  portMappings:
+  - 8080:8080
+routes:
+- to:
+    kind: Service
+    name: httpd
--- a/docs/file-reference.md
+++ b/docs/file-reference.md
@@ -423,6 +423,38 @@ More info about Probe: https://kubernetes.io/docs/api-reference/v1.6/#probe-v1-c
 
 The name of the Ingress.
 
+## routes
+
+```yaml
+routes:
+- <routeObject>
+- <routeObject>
+```
+
+| **Type**                                  | **Required** |
+|-------------------------------------------|--------------|
+| array of [route object](#routeObject) | no           |
+
+
+### routeObject
+
+```yaml
+routes:
+- name: <string>
+  <Route Spec>
+```
+
+Example:
+```yaml
+name: webroute
+to:
+  kind: Service
+  name: httpd
+```
+
+Each `route` object is OpenShift's `RouteSpec` and `ObjectMeta` field.
+More info: https://docs.openshift.org/latest/rest_api/apis-route.openshift.io/v1.Route.html#object-schema
+
 ## secrets
 
 ```yaml
--- a/pkg/spec/resources.go
+++ b/pkg/spec/resources.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
+	os_route_v1 "github.com/openshift/origin/pkg/route/apis/route/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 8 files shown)


containscafeine:
- [ ] add docs
- [ ] add tests
- [ ] add examples

pkg/spec/util.go
```diff
@@ -65,12 +66,20 @@ func GetScheme() (*runtime.Scheme, error) {
 	// Initializing the scheme with the core v1 api
 	scheme := api.Scheme
 
+	// TODO: find a way where we don't have to add all the subsequent schemes
```
kadel: Why different schema for different controllers? 
Using one with everything should be fine or not?

containscafeine: It's fine, but it'd lead to a _huge_ schema. So this is more of an optimization TODO, does not really matter much right matter tbh.

containscafeine:
ping @kadel @cdrage @surajssd - I've added docs, tests and an example.
PTAL!
I might not be online tomorrow and day after (19th, 20th October), so feel free to edit the PR (edits are allowed by maintainers)
I've also broken changes into different commits, so should be easier to review.
If you folks need to merge, use the "Squash and merge" button, just in case :tada:

docs/examples/routes/httpd.yml
```diff
@@ -0,0 +1,13 @@
+name: httpd
+containers:
+- image: centos/httpd
```
kadel: This image uses root,so it can't be run on default openShift configuration :(

kadel: but you could use  `bitnami/nginx` for exaple

docs/file-reference.md
```diff
@@ -423,6 +423,41 @@ More info about Probe: https://kubernetes.io/docs/api-reference/v1.6/#probe-v1-c
...
+```
+
+Example:
+```yaml
+name: webroute
+host: httpd-web.192.168.42.69.nip.io
+to:
+  kind: Service
+  name: httpd
+  weight: 100
```
kadel: I would remove `weight` and `wildcardPolicy` It just makes it look too complicated

docs/file-reference.md
```diff
@@ -423,6 +423,41 @@ More info about Probe: https://kubernetes.io/docs/api-reference/v1.6/#probe-v1-c
...
+```yaml
+routes:
+- name: <string>
+  <Route Spec>
+```
+
+Example:
+```yaml
+name: webroute
+host: httpd-web.192.168.42.69.nip.io
```
kadel: can we maybe remove `host`?
It is not required, and if someone tries it like this it won't work

pkg/spec/resources.go
```diff
@@ -225,6 +252,19 @@ func (app *ControllerFields) createIngresses() ([]runtime.Object, error) {
 	return ings, nil
 }
 
+func (app *ControllerFields) createRoutes() ([]runtime.Object, error) {
```
kadel: It would be nice to have tests for this function ;-)

kadel:
Just a few notes regarding docs, and test. The code itself looks good.

docs/examples/routes/httpd.yml
containscafeine: fixed

docs/file-reference.md
containscafeine: agree, removed

pkg/spec/resources.go
containscafeine: added

PR #386, opened 2017-10-26
few small updates to travis.yml

kadel:
This PR contains a couple of fixes for travis-ci. Two of them are one-liners, it doesn't make sense to create separate pr for every one of them.

- upload binaries only from master branch
- don't run go test in verbose mode (it just pollutes output with unnecessary information, if test fails it will show all the necessary info)
- generate and upload info.txt to bintray so it is possible to get information about what was uploaded


cdrage:
LGTM

PR #387, opened 2017-10-26
Sync docs with gh-pages

cdrage:
Adds introduction.md as well as syncs all the documentation from /docs
to gh-pages.


cdrage:
I'm going to merge this in and see if it correctly syncs!

PR #388, opened 2017-10-26
Remove verbose output e2e tests

cdrage:
Removes the verbose output of e2e tests and slightly changes the output
y adding an echo of using `kubectl get po --all-namespaces -w`.

### DIFF
--- a/scripts/run_e2e.sh
+++ b/scripts/run_e2e.sh
@@ -31,18 +31,36 @@ if [ -n "$TIMEOUT" ]; then
     run_command+=" -timeout=$TIMEOUT"
 fi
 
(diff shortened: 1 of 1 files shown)


ashetty1:
@cdrage I think another way to do this would be to add `run_e2e.sh -v' option to the script itself. WDYT?

Cc: @surajssd @kadel

cdrage:
@ashetty1 you mean `make test-e2e -v` ?

ashetty1:
@cdrage Yes. `make VERBOSE=yes / no`

cdrage:
Updated with your suggested changes @ashetty1 let me know if these are good.

PR #391, opened 2017-10-26
Add semaphore badge to README

cdrage:
Adds the Semaphore badge to README (now that Semaphore is stable enough
that it's not erroring out all the time).


kadel:
Something is wrong with it. Markdown is not rendering correctly 
:-(

containscafeine:
@kadel fixed, was a small type

PR #423, opened 2017-11-06
Update README.md and add installation guide

cdrage:
Simplifies the README.md a bit, since we not only reference user / quick
astart guides elsewhere, but some of the information is redundant.
Removes "project status".

An installation doc is also added too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,29 +2,24 @@
 
 [![Semaphore Build Status Widget]][Semaphore Build Status] [![Travis Build Status Widget]][Travis Build Status] [![Coverage Status Widget]][Coverage Status] [![GoDoc Widget]][GoDoc] [![GoReportCard Widget]][GoReportCardResult] [![Slack Widget]][Slack] 
 
-## What is Kedge?
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -2,31 +2,26 @@
...
-We are a very evolving project with high velocity, we have listed a [file reference specification](docs/file-reference.md) as well as document our RFC's and changes as [GitHub issues](https://github.com/kedgeproject/kedge/issues).
 
-Check out our [roadmap](ROADMAP.md) as we push towards a __0.4.0__ release.
+----
 
-## Using Kedge
+## To start using Kedge
 
-### Installing
+### Install
```
kadel: Can we link to installation.md somewhere in this section?
It includes some important information like latest builds, @pradeepto this information in readme

cdrage: I'll update it. Thanks!

cdrage: I put the latest builds in a separate installation document, but we could move this back to the README.md if you wish.

kadel: 👍 I'm fine with just linking it.

PR #424, opened 2017-11-06
Update file reference for deploymentconfig

cdrage:
Updates the file reference file with references to deploymentconfig

### DIFF
--- a/docs/file-reference.md
+++ b/docs/file-reference.md
@@ -161,30 +161,55 @@ name: mariadb
 controller: deployment
 ```
 
+> Example using DeploymentConfig for OpenShift
+
+```yaml
+controller: deploymentconfig
+name: httpd
+replicas: 2
+containers:
+- image: bitnami/nginx
+services:
+- name: httpd
+  type: NodePort
+  ports:
+  - port: 8080
+    targetPort: 8080
+```
+
+> Example using Job
+
+```yaml
+controller: job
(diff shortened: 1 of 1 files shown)


docs/file-reference.md
```diff
@@ -161,30 +161,55 @@ name: mariadb
...
 ```
 
+> Example using DeploymentConfig for OpenShift
+
+```yaml
+controller: deploymentconfig
+name: httpd
+replicas: 2
+containers:
+- image: centos/httpd
```
kadel: this container won't work on OpenShift  (root privileges)

kadel: you can use `bitnami/nginx` or `tomaskral/nonroot-nginx` (not sure if there is openshift's or redhat's official image)

cdrage: thanks. updated.

docs/file-reference.md
```diff
@@ -161,30 +161,55 @@ name: mariadb
...
+name: httpd
+replicas: 2
+containers:
+- image: bitnami/nginx
+services:
+- name: httpd
+  type: NodePort
+  ports:
+  - port: 8080
+    targetPort: 80
```
kadel: one more small change. `bitnami/nginx` is listening in `8080`

cdrage: done.

PR #437, opened 2017-11-07
Update README with concise sentences, update quickstart guide

cdrage:
This updates the quickstart guide as well as the README.md and
introduction.md

I found that the points kind-of repeat themselves (See Concise and
Simplicity points).

Instead, alternative points such as "shortcuts" and reducing the overall
file-size of your file has been added.

Also, emphasizing that you can still use Kubernetes constructs within
your Kedge file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,12 @@
 
 Kedge is a simple, easy and declarative way to define and deploy applications to Kubernetes by writing very concise application definitions.
 
-Key features and goals include:
+Why do people love Kedge?
 
-  - __An extension of Kubernetes:__ Use pre-existing definitions such as Pod or Container within your artifact file.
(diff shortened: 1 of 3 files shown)


pradeepto:
I am not sure of the phrase "extensions". Because aren't CRDs extensions of Kubernetes?

kadel:
> Because aren't CRDs extensions of Kubernetes?

Don't think so.

But I agree that extension "An extensions of Kubebernetes" might not a bit misleading.

"An extensions of Kubebernetes definitions" or something along those lines, something that explains that is extending kubernetes artifacts/definitions

cdrage:
@kadel @pradeepto Updated with your suggestions `an extension of Kubernetes definitions`

PR #461, opened 2017-11-17
This will add description about the new field appversion

piyush1594:
This will add description about the new field appversion
in file-reference.md
#407

### DIFF
--- a/docs/file-reference.md
+++ b/docs/file-reference.md
@@ -103,6 +103,7 @@ For example, Kedge simplifies deployment by introducing the `health` key. Howeve
 
 ```yaml
 name: <string>
+appversion: <string>
 controller: <string>
 labels: <object>
 containers:
@@ -137,6 +138,7 @@ Depending on the controller key selected. Each "app" (Kedge file) is an extensio
 | Field    | Type     | Required     | Description  |
 |----------|----------|--------------|--------------|
 | name | string   | yes          | The name of the app or microservice this particular file defines. |
+| appversion | string | no           | The version of the app or microservice this particular file defines. |
 | controller | string   | no           | The Kubernetes controller of the app or microservice this particular file (default: "deployment") |
 | labels | object   | no           | Map of string keys and values that can be used to organize and categorize (scope and select) objects. May match selectors of replication controllers and services. |
 | containers | array of [containerObject](#containerobject) | yes          | [containerObject](#containerobject) |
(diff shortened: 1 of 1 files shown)


docs/file-reference.md
```diff
@@ -1007,4 +1019,4 @@ services:
   ports:
   - port: 8080
     targetPort: 80
-```
\ No newline at end of file
+```
```
cdrage: please don't change this line of code.

docs/file-reference.md
```diff
@@ -155,6 +156,17 @@ name: mariadb
...
 
+## appversion
+
+```yaml
+appversion: "5.0.0-alpha"
+```
+
+| Type     | Required     | Description  |
+|----------|--------------|--------------|
+| string   | yes          | The version of the app or micro-service this particular file defines. |
```
cdrage: no, it's not required.

docs/file-reference.md
```diff
@@ -155,6 +156,17 @@ name: mariadb
 | string   | yes          | The name of the app or micro-service this particular file defines. |
 
 
+## appversion
+
+```yaml
+appversion: "5.0.0-alpha"
```
cdrage: honestly, i'd put something like 0.0.1 instead of putting it in quotes. looks better / much easier.

piyush1594: @cdrage I put it in quotes because, if someone is giving version as 5 or 5.1 then the user needs to give in quotes, it will work without quotes only in case of a string.
So I don't know what I should finally take as an example.

kadel: I would actually prefer putting it in quotes.

kadel: Let's use quotes for now, and than we can change is based on how #407 will be implemented

cdrage: I think it's ugly to have quotes... The power of YAML is that it can be either one when unmarshaling.

Based on when #455 is merged, we can make it whatever we want it to be (int or string) if someone were to put a float or a version value.

kadel: We can't merge this before #407 anyway, so let's wait for how it will be implemented.

kadel: @cdrage check https://github.com/kedgeproject/kedge/pull/454#discussion_r152602457

docs/file-reference.md
```diff
@@ -136,6 +137,7 @@ Each "app" (Kedge file) is a Kubernetes <a target="_blank" href="https://v1-6.do
 | Field    | Type     | Required     | Description  |
 |----------|----------|--------------|--------------|
 | name | string   | yes          | The name of the app or micro-service this particular file defines. |
+| appversion | string | no           | The version of the app or micro-service this particular file defines. |
```
cdrage: microservice is one word

piyush1594: hi @cdrage,

If you check name and controller field in the same table, it is written like this, if you still want the change then I will do.

Thanks

cdrage: yeah, both need to be changed :+1:

cdrage: pushed a PR here: https://github.com/kedgeproject/kedge/pull/493

cdrage:
LGTM

PR #467, opened 2017-11-20
add BuildConfig docs and example

containscafeine:
This commit adds the documentation and example for using
OpenShift BuildConfigs in Kedge

### DIFF
--- a/docs/examples/buildConfigs/README.md
+++ b/docs/examples/buildConfigs/README.md
@@ -1,0 +1,29 @@
+# BuildConfig
+
+A build configuration describes a single build definition and a set of triggers for when a new build should be created. A build configuration is defined by a BuildConfig.
+
+In Kedge, defining BuildConfigs is implemented by merging `BuildConfigSpec` and `ObjectMeta` at the root YAML level of the spec.
+
+A BuildConfig can be defined in Kedge as follows -
+
+```yaml
+buildConfigs:
+- triggers:
+  - type: "ImageChange"
+  source:
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+  strategy:
+    type: "Source"
+    sourceStrategy:
+      from:
+        kind: "ImageStreamTag"
+        name: "ruby-22-centos7:latest"
+  output:
+    to:
+      kind: "ImageStreamTag"
+      name: "origin-ruby-sample:latest"
+  postCommit:
+      script: "bundle exec rake test"
+```
--- a/docs/examples/buildConfigs/ruby-hello-world.yml
+++ b/docs/examples/buildConfigs/ruby-hello-world.yml
@@ -1,0 +1,43 @@
+controller: DeploymentConfig
+name: ruby
+containers:
+- image: ""
+triggers:
+- imageChangeParams:
+    automatic: true
+    containerNames:
+    - ruby
+    from:
+      kind: ImageStreamTag
+      name: origin-ruby-sample:latest
+  type: ImageChange
+services:
+- portMappings:
+  - "8080"
+imageStreams:
+- name: origin-ruby-sample
+- name: ruby-22-centos7
+  tags:
+  - name: latest
+    from:
+      kind: DockerImage
+      name: openshift/ruby-20-centos7
+buildConfigs:
+- triggers:
+  - type: "ImageChange"
+  source:
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+  strategy:
+    type: "Source"
+    sourceStrategy:
+      from:
+        kind: "ImageStreamTag"
+        name: "ruby-22-centos7:latest"
+  output:
+    to:
+      kind: "ImageStreamTag"
+      name: "origin-ruby-sample:latest"
+  postCommit:
+      script: "bundle exec rake test"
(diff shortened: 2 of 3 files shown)


docs/examples/buildConfigs/README.md
```diff
@@ -0,0 +1,28 @@
+# BuildConfig
+
+A build configuration describes a single build definition and a set of triggers for when a new build should be created. A build configuration is defined by a BuildConfig.
+
+In Kedge, defining BuildConfigs has been implemented by merging `BuildConfigSpec` and `ObjectMeta` at the root YAML level of the spec.
```
cdrage: `defining BuildConfigs are implemented`

docs/examples/buildConfigs/README.md
```diff
@@ -0,0 +1,28 @@
...
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+    dockerfile: "FROM openshift/ruby-22-centos7\nUSER example"
+  strategy:
+    type: "Source"
+    sourceStrategy:
+      from:
+        kind: "ImageStreamTag"
+        name: "ruby-20-centos7:latest"
```
kadel: should be ruby-22-centos7

docs/examples/buildConfigs/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```yaml
+buildConfigs:
+- triggers:
+  - type: "ImageChange"
+  source:
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+    dockerfile: "FROM openshift/ruby-22-centos7\nUSER example"
```
kadel: how is this going to work, when there is `docekrfile` is specified, but build strategy is `sourceStrategy` (S2I)?

This doesn't look good to me.

kadel: I've tried that and `dockerfile` is being ignored

docs/examples/buildConfigs/ruby-hello-world.yml
```diff
@@ -0,0 +1,44 @@
...
+      kind: DockerImage
+      name: openshift/ruby-20-centos7
+buildConfigs:
+- triggers:
+  - type: "ImageChange"
+  source:
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+    dockerfile: "FROM openshift/ruby-22-centos7\nUSER example"
```
kadel: again this doesn't seem right :-(

docs/file-reference.md
```diff
@@ -883,6 +895,73 @@ secrets:
...
+```yaml
+name: ruby
+buildConfigs:
+- triggers:
+  - type: "ImageChange"
+  source:
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+...
```
kadel: it would be better to show functional buildConfig here or at least one that can be created. This is just portion of it that can't be even created because it doesn't have all required fields

docs/file-reference.md
```diff
@@ -883,6 +895,73 @@ secrets:
...
+```yaml
+name: ruby
+buildConfigs:
+- triggers:
+  - type: "ImageChange"
+  source:
+    type: "Git"
+    git:
+      uri: "https://github.com/openshift/ruby-hello-world"
+    dockerfile: "FROM openshift/ruby-22-centos7\nUSER example"
```
kadel: again, dockerfile with soutrceStrategy, I don't think this is right

docs/examples/buildConfigs/README.md
containscafeine:
```console
$ oc explain bc.spec.source.dockerfile
FIELD: dockerfile <string>

DESCRIPTION:
     dockerfile is the raw contents of a Dockerfile which should be built. When
     this option is specified, the FROM may be modified based on your strategy
     base image and additional ENV stanzas from your strategy environment will be
     added after the FROM, but before the rest of your Dockerfile stanzas. The
     Dockerfile source type may be used with other options like git - in those
     cases the Git repo will have any innate Dockerfile replaced in the context
     dir.
```

docs/examples/buildConfigs/ruby-hello-world.yml
containscafeine: @kadel posted above, this is allowed with git.

docs/examples/buildConfigs/README.md
kadel: I saw that @containscafeine  ;-), But don't understand how this can work with S2I. S2I is not using Dockerfile does it?

docs/examples/buildConfigs/ruby-hello-world.yml
containscafeine: I think we can remove this from this example to make it more concise maybe

docs/examples/buildConfigs/README.md
containscafeine: should it not be - `defining BuildConfigs is implemented`?

docs/examples/buildConfigs/README.md
containscafeine: changed

containscafeine: changed

docs/file-reference.md
containscafeine: done

containscafeine:
@cdrage @kadel made the changes, PTAL!

cdrage:
LGTM

PR #497, opened 2017-11-28
Fixed kedge build error(archive/tar: write too long)

### LINKED ISSUE #490: kedge build returns - unable to create a tarball: archive/tar: write too long
My current directly size is -
```console
$ du -sh
812M    .
```

And I have simple Dockerfile in the current directory -
```console
$ cat Dockerfile 
FROM centos:7
CMD ping
```

Upon running `kedge build -i imagename` I expect a simple image to be built, instead I get the following error -
```console
$ kedge build -i testrun
INFO[0000] Building image 'testrun' from directory 'kedge' 
unable to create a tarball: archive/tar: write too long
```

This error does not occur when the current directory size is small.
I don't think so building an image should be dependent on the size of the current directory.

surajnarwade:
Fixes #490

Current directory size:

```
$ du -sh
799M	.
```

Dockerfile:

```
$ cat Dockerfile
FROM centos:7
CMD ping
```

Upon running above Dockerfile, Now kedge can successfully built an image.

```
$ kedge build -i testrun
INFO[0000] Building image 'testrun' from directory 'kedge'
INFO[0010] Image 'testrun' from directory 'kedge' built successfully
```

(Problem resolved by adding check for file mode bits)

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -153,21 +153,34 @@ func CreateTarball(source, target string) error {
 			if err != nil {
 				return err
 			}
-			header, err := tar.FileInfoHeader(info, info.Name())
+			var symLink string
+			// Checking for the symlinks
+			if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+				// Allow symlinks
+				if symLink, err = os.Readlink(path); err != nil {
+					return err
+				}
+			}
+
+			fileHeader, err := tar.FileInfoHeader(info, symLink)
 			if err != nil {
 				return err
 			}
 
 			if baseDir != "" {
 				if strings.HasSuffix(source, "/") {
-					header.Name = strings.TrimPrefix(path, source)
+					fileHeader.Name = strings.TrimPrefix(path, source)
 				} else {
-					header.Name = filepath.Join(baseDir, strings.TrimPrefix(path, source))
+					fileHeader.Name = filepath.Join(baseDir, strings.TrimPrefix(path, source))
 				}
 			}
 
-			if err := tarball.WriteHeader(header); err != nil {
+			if err := tarball.WriteHeader(fileHeader); err != nil {
 				return err
+			}
+
+			if !info.Mode().IsRegular() {
+				return nil
 			}
 
 			if info.IsDir() {
--- a/tests/cmd/cmd_test.go
+++ b/tests/cmd/cmd_test.go
@@ -15,6 +15,8 @@ import (
 var Fixtures = os.ExpandEnv("$GOPATH/src/github.com/kedgeproject/kedge/tests/cmd/fixtures/")
 var ProjectPath = "$GOPATH/src/github.com/kedgeproject/kedge/"
 var BinaryLocation = os.ExpandEnv(ProjectPath + "kedge")
+var imagename = "testrun"
+var context = "kedge-build/"
 
 func TestKedgeGenerate(t *testing.T) {
 	testCases := []struct {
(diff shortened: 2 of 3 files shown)


kadel:
This doesn't solve problem. Effect of this will be that any symlinks in the directory will be skipped.

We need proper fix that will allow symlinks to work.

surajnarwade:
@kadel Can you please review :)

kadel:
This still won't work in the same way Docker does it.
Symlinks are not handled correctly.
They should be converted to files inside tar, at least this is what docker does.

I've created simple test where we can try it.: https://github.com/kadel/build-with-symlink
```
▶ docker build .
Sending build context to Docker daemon  56.83kB
Step 1/4 : FROM busybox
 ---> 00f017a8c2a6
Step 2/4 : COPY link /link
 ---> Using cache
 ---> b21e58349d4d
Step 3/4 : RUN cat /link
 ---> Using cache
 ---> 4cf65dce686f
Step 4/4 : RUN ls -lah /link
 ---> Running in 86d18a0ba4c0
-rw-r--r--    1 root     root           5 Nov 30 11:38 /link
 ---> a8da85d6820c
Removing intermediate container 86d18a0ba4c0
Successfully built a8da85d6820c


▶ kedge build -i test .
INFO[0000] Building image 'test' from directory 'build'
unable to build image: COPY failed: Forbidden path outside the build context: link ()
```

kadel:
something like this might help https://stackoverflow.com/questions/38454850/getting-write-too-long-error-when-trying-to-create-tar-gz-file-from-file-and-d

kadel:
It might be worth to create some unit test for this (symlinks in build directory)

surajnarwade:
@kadel can you please test it now ?

pkg/build/build.go
```diff
@@ -143,7 +143,15 @@ func CreateTarball(source, target string) error {
 			if err != nil {
 				return err
 			}
-			header, err := tar.FileInfoHeader(info, info.Name())
+			var link string
```
cdrage: not a great name, symLink maybe?

pkg/build/build.go
```diff
@@ -143,7 +143,15 @@ func CreateTarball(source, target string) error {
...
+			if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+
+				if link, err = os.Readlink(path); err != nil {
```
cdrage: nested if statement, may be better with `if info.Mode()&os.ModeSymlink == os.ModeSymlink &&  link, err = os.Readlink(path); err != nil`

pkg/build/build.go
```diff
@@ -143,7 +143,15 @@ func CreateTarball(source, target string) error {
...
+					return err
+				}
+			}
+
+			header, err := tar.FileInfoHeader(info, link)
```
cdrage: what is 'header'? maybe, fileHeader would be better?

pkg/build/build.go
surajnarwade: didn't work :(

surajnarwade: @cdrage I got following error, 
```
pkg/build/build.go:148:52: invalid operation: info.Mode() & os.ModeSymlink == os.ModeSymlink && symLink (mismatched types bool and string)
```

cdrage: Post the full code.  This doesn't look like the full-line if-statement messed it up, but rather the `info.Mode()&os.ModeSymlink == os.ModeSymlink` part.

kadel:
The code looks good and as far as I tested it solves the problem.
It would be great to have some tests for this. So we can be sure that we don't break it accidentally in the future.

cdrage:
Tests pass, code LGTM.

Issue #491, opened 2017-11-27
Specifying default value for Kedge variable


kadel:
Allow specifying a default value for a variable if it is not set.

example kedge file:
```yaml
name: foo
image: foo/bar:[[ TAG:latest ]]
```

If `TAG` variable is not set, it `latest` will be used.

The rule for detecting default value should be: use everything between `:` and `]]` and trim whitespaces from both ends.

surajnarwade:
I will pick this up

PR #535, opened 2017-12-18
Specify Default value for variable

surajnarwade:
Solves #491

Now we can mention default value for variable. for example,

```
name: foo
image: foo/bar:[[ TAG:latest ]]
```

First Kedge will look value of `TAG` in environment and if there's no any value in environment, it will
give default value.

### DIFF
--- a/docs/file-reference.md
+++ b/docs/file-reference.md
@@ -1067,6 +1067,18 @@ NGINX_VERSION=1.13 kedge apply -f nginx.yaml
 
 You can use variables anywhere in the Kedge file. Variable names are enclosed in double square brackets (`[[ variable_name ]]`). For example `[[ IMAGE_NAME ]]` will be replaced with value of environment variable `$IMAGE_NAME`.
 
+> Specify default value of variable,
+
+```yaml
+name: foo
+image: foo/bar:[[ TAG:latest ]]
+```
+
+You can specify a default value of a variable if it is not set. 
+For example: `[[ TAG:latest ]]`
+If `TAG` variable is not set, `latest` will be used.
+
+
 # Controllers
 
 There are three defineable controllers within Kedge:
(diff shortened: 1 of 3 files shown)


pkg/cmd/util_test.go
```diff
@@ -188,6 +188,32 @@ func TestSubstituteVariables(t *testing.T) {
 			[]byte(""),
 			true,
 		},
+		{
+			[]byte(`
+				name: httpd
+				containers:
+				- image: foo/bar:[[ TEST_IMAGE_TAG:latest ]]
```
kadel: can we have tests for following cases?:

`[[ TEST_IMAGE_TAG: latest ]]`
`[[ TEST_IMAGE_TAG: latest]]`
`[[TEST_IMAGE_TAG:latest]]`

surajnarwade: sure

surajnarwade: @kadel , added test for above cases

kadel:
Code looks good, we just need some documentation, and its done

docs/file-reference.md
```diff
@@ -1067,6 +1067,16 @@ NGINX_VERSION=1.13 kedge apply -f nginx.yaml
 
 You can use variables anywhere in the Kedge file. Variable names are enclosed in double square brackets (`[[ variable_name ]]`). For example `[[ IMAGE_NAME ]]` will be replaced with value of environment variable `$IMAGE_NAME`.
 
+You can specify default value of variable if it's not set. for example, `[[ TAG:latest ]]`
```
cdrage: You need to move these two sentences BELOW the example (after the yaml code) or else the formatting is a bit messed up.

cdrage: `You can specify a default value of a variable if it is not set.`

cdrage: `For example: `[[ TAG:latest ]]`

cdrage:
LGTM :+1:

PR #544, opened 2017-12-26
Making size field optional in PVC

surajnarwade:
Added `//+optional` for `size` key in `VolumeClaim` defination.
This will make this field option in JSON schema.

### DIFF
--- a/pkg/spec/types.go
+++ b/pkg/spec/types.go
@@ -35,6 +35,7 @@ type VolumeClaim struct {
 	// k8s: io.k8s.kubernetes.pkg.api.v1.PersistentVolumeClaimSpec
 	api_v1.PersistentVolumeClaimSpec `json:",inline"`
 	// Size of persistent volume
+	// +optional
 	Size string `json:"size"`
 	// k8s: io.k8s.kubernetes.pkg.apis.meta.v1.ObjectMeta
 	meta_v1.ObjectMeta `json:",inline"`


pkg/spec/types.go
```diff
@@ -35,6 +35,7 @@ type VolumeClaim struct {
 	// k8s: io.k8s.kubernetes.pkg.api.v1.PersistentVolumeClaimSpec
 	api_v1.PersistentVolumeClaimSpec `json:",inline"`
 	// Size of persistent volume
+	// +optional
```
kadel: Size is a required field! PersistentVolumeClaim can't be created without size.

surajnarwade: As per file reference, `size` conflicts with  `resources`, Hence `size` can be optional, I think so

surajnarwade: Reference: http://kedgeproject.org/file-reference/#persistentvolumeobject

surajnarwade:
@kadel , problem is here,

```
$ kedge generate -f allnomagic/db.yaml --enable-validation 
The kedgefile is not valid. see errors :
- size: size is required
```

pkg/spec/types.go
kadel: That is true, but one of them has to be set, either 
```
  resources:
    requests:
      storage: 500Mi
```
or 
`size: 500Mi`

kadel:
Ok, let's go with this for now. 
I've opened https://github.com/kedgeproject/kedge/issues/562 so we can fix this properly later.

pkg/spec/types.go
kadel: yes and no.
Exactly one has to be set either `resources.requests.storage` or `size.
One of them has to be set otherwise, PVC can't be created.

One of those has to be set. If both are set than it is conflict and we don't know which one to use. And if neither is set than its incomplete PVC definition.

PR #550, opened 2018-01-02
Remove test-jsonschema-generation make target

concaf:
This commit removes the test-jsonschema-generation make target
since the older image is no longet maintained and the newer
types.go will be checked during the travis run


containscafeine:
@cdrage @kadel can we merge this since this is increasing the CI time by _a lot_ and the image is no longer maintainer and up to date.

kadel:
can we change so it can be used to locally generate schema using new image?

PR #553, opened 2018-01-02
Pass absolute path to JSON Schema docker command

containscafeine:
To mount a host directory, we need to use absolute path to the
JSON Schema container image. Before this commit, the mounted path
was relative and was not producing the desired output. This commit
fixes that.


containscafeine:
Merging to test fixes

PR #563, opened 2018-01-05
Widget image no longer active

cdrage:
Widget image for slack no longer works. Removing the widget as well as
the link.


kadel:
I agree we can remove it, as we have a link to slack below in community section in readme.

Issue #560, opened 2018-01-03
Lowercase "tcp" or "udp" does not work for portMappings.


cdrage:
See error:

```sh
Unable to perform controller operations: unable to transform data: failed to create Kubernetes objects: Unable to create Kubernetes Service: unable to parse port mapping: invalid protocol 'tcp' provided, the acceptable values are 'TCP' and 'UDP'
```

When using:

```yaml
portMappings:
 - 8080:8080/tcp
```

We should support both upper and lowercase.

kadel:
agreed, this should be case insensitive

cdrage:
#564 has been merged. closing!

PR #564, opened 2018-01-06
Added case for lowercase protocol

piyush1594:
This will accept if someone specifies protocol as
tcp or Tcp or tCp. Likewise UDP also
Always return the protocol as Uppercase
Added test for two cases
#560

### DIFF
--- a/pkg/spec/resources.go
+++ b/pkg/spec/resources.go
@@ -669,9 +669,10 @@ func parsePortMapping(pm string) (*api_v1.ServicePort, error) {
 	// Case 3 - port/protocol
 	// Case 4 - port:targetPort/protocol
 	case 2:
-		switch api_v1.Protocol(protocolSplit[1]) {
+		protocolUpperCase := strings.ToUpper(protocolSplit[1])
+		switch api_v1.Protocol(protocolUpperCase) {
 		case api_v1.ProtocolTCP, api_v1.ProtocolUDP:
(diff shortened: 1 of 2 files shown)


pradeepto:
Thank you @piyush1594 for fixing issues in the Open Source spirit.

pkg/spec/resources.go
```diff
@@ -669,9 +669,10 @@ func parsePortMapping(pm string) (*api_v1.ServicePort, error) {
 	// Case 3 - port/protocol
 	// Case 4 - port:targetPort/protocol
 	case 2:
-		switch api_v1.Protocol(protocolSplit[1]) {
+		protocol_uppercase := strings.ToUpper(protocolSplit[1])
```
kadel: just one small nitpick :innocent: 

 `protocolUpperCase`

Go formatting style prefers camelCase names.
Yes, I know we have other variables with an underscore like `api_v1` :smile: , but that is a special case.
Normal variables should be camelCase

cdrage:
Code LGTM. Same as @kadel just change to camelCase and let's merge :+1:

surajnarwade:
LGTM (after camelCase)

piyush1594:
Changes requested by you are done. Please take a look.

Thanks, @pradeepto @kadel @cdrage @surajnarwade

cdrage:
LGTM! Thank you :+1:

PR #565, opened 2018-01-09
New Kedge file structure

### LINKED ISSUE #541: Changing structure? (Kedge v2)  - Creating controllers section
We are starting to see some cases that are complicated because of our decisions to have controller objects on the root level.

- #540 is one of the cases,
- other is JSON schema with our current design it is not possible to describe Kedge spec with one JSON schema file. This will complicate doing LSP for editors.
- we are also putting and treating controller objects as a special object that is more important than others 

I was thinking if it would make sense to change the structure a bit, and put controllers in separate sections the same way we have `services`, `ingresses`, `configMaps`....., This will also make easier to distinguish between global and local (object specific) values.

Generic structure would look like this:

```
<ObjectMeta>
appVersion

<kind>:
   - <ObjectMeta> merged with <kindSpec>
```

Root level `ObjectMeta` will be always merged with local `ObjectMeta` for given object.
That way we can define the common name, global labels and annotations.

It would look like this:

```yaml
name: httpd
appVersion: 0.1

deployments:
- containers:
  - image: centos/httpd

services:
- type: LoadBalancer
  portMappings: 
   - 8080:80
```
Benefits:
 - clear boundaries between controller Kubernetes object and Kedge keys like appVersion (no need to mix keys like appVersion into Kubernetes controller)
 - one generic structure, no object has a special place - easier for users - better UX)
 - allows multiple controllers in one file (we would no longer need `---` "hack") - better UX
 - easier to handle global labels, annotations, namespace etc... - easier to implement, better UX
 - can be described by one JSON Schema file, easier handling and generation,  easier to implement smart completion for editors.

_We should also introduce `formatVersion`, or `kedgeVersion` field to allow kedge spec versioning._

kadel:
fixes #541

first version - Kubernetes deployment should work

TODO:
 - [x] fix unit tests
 - [x] fix cmd tests
 - [x] update rest of the examples
 - [x] add Jobs back
 - [x] fix DeploymentConfigs
 - [ ] update documentation (will be part of the different PR) 

  

issues:
 - [x] Deployment name is missing
 - [x] template selector is missing (verify this)

my notes :
  -  what is a difference between fix* and populate* functions? 
  - `validateVolumeClaims` ?
 - `fixControllerFields` ?
 - a lot of code duplication between deployment and deploymentConfig

### DIFF
--- a/docs/examples/all/db.yaml
+++ b/docs/examples/all/db.yaml
@@ -1,33 +1,34 @@
 name: database
-containers:
-- image: mariadb:10
-  env:
-  - name: MYSQL_ROOT_PASSWORD
-    value: rootpasswd
-  - name: MYSQL_PASSWORD
-    value: wordpress
-  - name: MYSQL_USER
-    value: wordpress
-  envFrom:
-  - configMapRef:
-      name: database
-  volumeMounts:
-  - name: database
-    mountPath: /var/lib/mysql
-  livenessProbe:
-    exec:
-      command:
-      - mysqladmin
-      - ping
-    initialDelaySeconds: 30
-    timeoutSeconds: 5
-  readinessProbe:
-    exec:
-      command:
-      - mysqladmin
-      - ping
-    initialDelaySeconds: 5
-    timeoutSeconds: 1
+deployments:
+- containers:
+  - image: mariadb:10
+    env:
+    - name: MYSQL_ROOT_PASSWORD
+      value: rootpasswd
+    - name: MYSQL_PASSWORD
+      value: wordpress
+    - name: MYSQL_USER
+      value: wordpress
+    envFrom:
+    - configMapRef:
+        name: database
+    volumeMounts:
+    - name: database
+      mountPath: /var/lib/mysql
+    livenessProbe:
+      exec:
+        command:
+        - mysqladmin
+        - ping
+      initialDelaySeconds: 30
+      timeoutSeconds: 5
+    readinessProbe:
+      exec:
+        command:
+        - mysqladmin
+        - ping
+      initialDelaySeconds: 5
+      timeoutSeconds: 1
 services:
 - name: database
   ports:
--- a/examples/guestbook-demo/db.yaml
+++ b/examples/guestbook-demo/db.yaml
@@ -1,18 +1,20 @@
 name: mongodb
 
-containers:
-  - image:  centos/mongodb-32-centos7:3.2
-    envFrom:
-      - secretRef:
-          name: mongodb-admin
-      - secretRef:
-          name: mongodb-user
-      - configMapRef:
-          name: mongodb-user
+deployments:
+- containers:
+    - image:  centos/mongodb-32-centos7:3.2
+      envFrom:
+        - secretRef:
+            name: mongodb-admin
+        - secretRef:
+            name: mongodb-user
+        - configMapRef:
+            name: mongodb-user
 
-    volumeMounts:
-     - name: mongodb-data
-       mountPath: /var/lib/mongodb/data
+      volumeMounts:
+       - name: mongodb-data
+         mountPath: /var/lib/mongodb/data
+
 services:
   - portMappings:
     - 27017
--- a/pkg/spec/controller_test.go
+++ b/pkg/spec/controller_test.go
@@ -24,42 +24,52 @@ import (
 	api_v1 "k8s.io/kubernetes/pkg/api/v1"
 )
 
-func TestUnmarshalValidateFixControllerOperations(t *testing.T) {
+func TestUnmarshalValidateFixApp(t *testing.T) {
 	tests := []struct {
 		Name string
 		Data []byte
-		App  *DeploymentSpecMod
+		App  *App
 	}{
 		{
 			Name: "One container mentioned in the spec",
 			Data: []byte(`
 name: test
-containers:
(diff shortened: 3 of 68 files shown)


kadel:
ping @containscafeine  @cdrage @surajnarwade 

can you please check new [types.go]( https://github.com/kedgeproject/kedge/pull/565/commits/d130e3d92dcdbc87d8cd83bba53b6fd1b0b880b6)?

I would like to verify that I'm on the good track with this before I start moving more code.

surajnarwade:
@kadel, `types.go` looks good to me

pkg/spec/controller_test.go
```diff
@@ -24,216 +24,232 @@ import (
...
-			continue
-		}
-
-		if reflect.TypeOf(kController) != reflect.TypeOf(test.ExpectedController) {
-			t.Errorf("Got wrong controller %#v. Expected: %#v. For data:\n%s\n", reflect.TypeOf(kController).String(), reflect.TypeOf(test.ExpectedController).String(), string(test.Data))
-		}
-
-	}
-}
+// TODO: v2 do we need this test?
```
cdrage: Probably not.

cdrage:
I did a quick look-over at the code, it's LGTM.

docs/examples/all/db.yaml
```diff
@@ -1,32 +1,33 @@
...
-    initialDelaySeconds: 30
-    timeoutSeconds: 5
-  readinessProbe:
-    exec:
-      command:
-      - mysqladmin
-      - ping
-    initialDelaySeconds: 5
+deployments:
+- containers:
```
cdrage: I'm just going through all these examples, but does `- containers` need to be an array? Can't it just be `containers:` ?@kadel

examples/guestbook-demo/db.yaml
```diff
@@ -1,18 +1,20 @@
...
-  - image:  centos/mongodb-32-centos7:3.2
-    envFrom:
-      - secretRef:
-          name: mongodb-admin
-      - secretRef:
-          name: mongodb-user
-      - configMapRef:
-          name: mongodb-user
+deployments:
+- containers:
```
cdrage: yeah.. looking at previous examples, I believe it should not be an array, just `containers`...

docs/examples/all/db.yaml
cdrage: Emphasized this here: https://github.com/kedgeproject/kedge/issues/541

kadel: Deployemnts should be an array like all the other (`services`, `configMaps` ....). It should be possible to define multiple deployments, and

examples/guestbook-demo/db.yaml
kadel: this should be an array. The one where it is not is wrong. (my mistake)

It is quite possible there will be more stuff that I might miss, it was a lot of examples to update.

kadel:
Test Fixed, (Kubernetes e2e on SemaphoreCI are OK, so does TravisCI tests)

kadel:
Ready for review, @cdrage @surajnarwade @containscafeine.

Please review this carefully. It involved a lot of code moving. Tests are passing, but it doesn't mean that I didn't make any mistakes. There is a big likelihood that I have missed something.

pkg/spec/controller_test.go
```diff
@@ -24,216 +24,232 @@ import (
...
+
+//func TestGetController(t *testing.T) {
```
cdrage: I think this could be totally removed.

kadel:
jsonSchema for testing: https://gist.github.com/kadel/25d8a188a80158185f25b5c6b76c50eb

pkg/spec/types.go
```diff
@@ -211,42 +223,44 @@ type ControllerFields struct {
...
 }
 
-// JobSpecMod is Kedge's extension of Kubernetes JobSpec and allows
-// defining a complete kedge application
+// JobSpecMod is Kedge's extension of Kubernetes JobSpec
 // kedgeSpec: io.kedge.JobSpecMod
 type JobSpecMod struct {
-	ControllerFields `json:",inline"`
+	// k8s: io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta
+	meta_v1.ObjectMeta `json:",inline88891"`
```
surajnarwade: /me confused here between `inline` & `inline88891`

surajnarwade:
Hey @kadel , I tried all examples and worked well and code also LGTM

pkg/spec/deployment.go
```diff
@@ -21,135 +21,122 @@ import (
...
-	return nil
-}
-
-func (deployment *DeploymentSpecMod) Validate() error {
-
-	// validate controller fields
-	if err := deployment.ControllerFields.validateControllerFields(); err != nil {
-		return errors.Wrap(err, "unable to validate controller fields")
-	}
+func (app *App) validateDeployments() error {
```
surajnarwade: is it related to #481 ?

kadel: nope, that is something different,

pkg/spec/types.go
kadel: ah ;-) that is typo :-D (psst yubi key :innocent: ;-) )

good catch

cdrage:
LGTM :+1:

Issue #558, opened 2018-01-03
Test all examples in docs/examples and examples


kadel:
We should make sure that all examples in `docs/examples` and in `examples` can be converted.

We have framework for command line tests (see https://github.com/kedgeproject/kedge/pull/498)

We should add test that will list all YAML files in directories with examples and run `kedge generate` on them.
This will be especially useful for #481

kadel:
related to https://github.com/kedgeproject/json-schema-generator/issues/25

kadel:
we can implement just https://github.com/kedgeproject/json-schema-generator/issues/25 as it is basically the same thing

surajnarwade:
I have implemented this test [here](https://github.com/kedgeproject/kedge/pull/481/commits/f78c5a4ef0b81c473c580be0762a57b342a29c6c#diff-2fd956fb798d519aa017383f8d85c53f) in PR #481

surajnarwade:
Closing this since #481 is merged now and we are testing all examples from `docs/examples` and `examples` here https://github.com/kedgeproject/kedge/blob/master/tests/cmd/cmd_test.go#L136

PR #566, opened 2018-01-10
Test kedgefiles from examples

surajnarwade:
This PR will test examples from `examples/` and `docs/examples` directory.

### DIFF
--- a/tests/cmd/cmd_test.go
+++ b/tests/cmd/cmd_test.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"testing"
 
+	"path/filepath"
+
 	"github.com/kylelemons/godebug/diff"
 )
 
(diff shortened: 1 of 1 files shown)


surajnarwade:
@kadel @cdrage review needed :)

surajnarwade:
Once https://github.com/kedgeproject/kedge/pull/565 gets merged, I have to change files accordingly

kadel:
why copy all examples to fixtures and duplicate it?

We should test them in `docs/examples`. One of the goals of this is to make sure that all our examples that users can access are valid.  But if we move them to fixtures they won't see it.

kadel:
Olso if you do it the right way that there will be no dependency on #565

surajnarwade:
@kadel files are not copied to fixtures, they are output files which will be used to verify.

kadel:
Ah sorry, :-( haven't looked properly. Sorry about that, my bad.

I was imagining something else, to be honest. This is all nice, but it will be hell to maintain it. 

If this is supposed to solve https://github.com/kedgeproject/kedge/issues/558 
than it just needs to verify that `kedge generate` doesn't throw an error.
The important thing here is that it does "autodiscovery" new examples without the need to add it to the test manually.

kadel:
This PR will serve as  main tests for https://github.com/kedgeproject/kedge/pull/481

tests/cmd/cmd_test.go
```diff
@@ -114,3 +115,37 @@ func Test_builderror(t *testing.T) {
...
+	}
+
+	// Test Examples from `docs/examples/ directory`
+	files, err = ioutil.ReadDir("../../docs/examples")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for _, f := range files {
+		if f.Name() == "README.md" || f.Name() == "build" || f.Name() == "s2i" {
```
kadel: Why is whole s2i dir skipped here? There kedge files in it which should be tested.

It might be better to just include all `*.yml` and `*.yaml` files instead of excluding the others.


Also, it might be better to use [filepath.Walk](
https://golang.org/pkg/path/filepath/#Walk) to get all the files (even those nested deep in the file structure)

tests/cmd/cmd_test.go
```diff
@@ -114,3 +116,25 @@ func Test_builderror(t *testing.T) {
 		t.Errorf("Test Failed")
 	}
 }
+
+func Test_examples(t *testing.T) {
+	searchDir := os.ExpandEnv(ProjectPath)
+	fileList := []string{}
+	err := filepath.Walk(searchDir, func(path string, f os.FileInfo, err error) error {
```
kadel: This should list only YAML files from `docs/examples` and from `examples`.

Sorry about that this is my fault, this wasn't clear from my previous comment :-(

PR #571, opened 2018-01-17
Add comparison documentation for Kedge

cdrage:
This adds a comparison table against similar projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,17 @@ Go through our [file reference](http://kedgeproject.org/file-reference).
 
 Then go further with our pre-existing [examples](https://github.com/kedgeproject/kedge/tree/master/examples).
 
+__Comparison to similar projects:__
+
+|                                          | Kompose | Ksonnet | OpenShift templates | Helm | Kedge |
+|------------------------------------------|---------|---------|---------------------|------|-------|
+| Scalable                                 | -       | ✓       | ✓                   | ✓    | ✓     |
+| Kubernetes compatible API and DSL Syntax | -       | -       | -                   | -    | ✓     |
+| Native to Kubernetes ecosystem           | -       | ✓       | -                   | ✓    | ✓     |
+| Integration to tools (IDEs)              | ✓       | ✓       | ✓                   | ✓    | ✓     |
+| YAML / JSON                              | ✓       | -       | ✓                   | ✓    | ✓     |
+| Parameterized                            | ✓       | ✓       | ✓                   | ✓    | ✓     |
+
 ## Community, Discussion, Contribution, and Support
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,6 +50,17 @@ Go through our [file reference](http://kedgeproject.org/file-reference).
 
 Then go further with our pre-existing [examples](https://github.com/kedgeproject/kedge/tree/master/examples).
 
+__Comparison to similar projects:__
+
+|                                          | Kompose | Ksonnet | OpenShift templates | Helm | Kedge |
+|------------------------------------------|---------|---------|---------------------|------|-------|
+| Scaleable                                | -       | ✓       | ✓                   | ✓    | ✓     |
```
surajnarwade: `Scalable`

cdrage:
Thanks @surajnarwade , updated :)

containscafeine:
Umm, this shows Kedge as a competitor to the other tools, while it does not necessarily overlap with Helm, Templates, etc.

README.md
```diff
@@ -50,6 +50,17 @@ Go through our [file reference](http://kedgeproject.org/file-reference).
...
 Then go further with our pre-existing [examples](https://github.com/kedgeproject/kedge/tree/master/examples).
 
+__Comparison to similar projects:__
+
+|                                          | Kompose | Ksonnet | OpenShift templates | Helm | Kedge |
+|------------------------------------------|---------|---------|---------------------|------|-------|
+| Scalable                                 | -       | ✓       | ✓                   | ✓    | ✓     |
+| Kubernetes compatible API and DSL Syntax | -       | -       | -                   | -    | ✓     |
+| Native to Kubernetes ecosystem           | -       | ✓       | -                   | ✓    | ✓     |
+| Integration to tools (IDEs)              | ✓       | ✓       | ✓                   | ✓    | ✓     |
```
surajnarwade: I am bit confused here for kompose

kadel: what @cdrage probably means is that basicaly every IDE that supporsts docker-compose has by definition support for kompose :-)

kadel (approved):
looks good. I would just consider explaining what Scalable means. For someone it might not be clear what is meant by that

concaf (changes requested):
I'm nack for this :(
This gives a false impression that kedge is comparable or is in the same domain and space as Helm or OpenShift templates, and will add to the confusion.

kadel:
> I'm nack for this :(
> This gives a false impression that kedge is comparable or is in the same domain and space as Helm or OpenShift templates, and will add to the confusion.

Almost everyone who sees Kedge for the first time asks quesion how Kedge compares (or how it is different) from those projects.

**We have answer this**.  
How do you suggest to explain it? It has to be in some short and easy to understand form.

containscafeine:
Agree, we should address this, but not by comparing. We can have something like https://www.nomadproject.io/intro/vs/index.html

Issue #555, opened 2018-01-03
Incorrect PR number in commit message


surajnarwade:
PR number link is incorrect in commit message to `json-schema` repo which is auto commited for every PR merge for this repo.

```
Commit: kedgeproject/kedge@6c61d45

PR: https://github.com/kedgeproject/kedge/pull/false
```

containscafeine:
Yeah, we should probably remove it since this is only put when the PR has already been merged i.e. it runs only on master. Hence, no PR number.

PR #573, opened 2018-01-17
Updates commit message for json-schema push

cdrage:
Removes the PR number in the commit message.

Closes https://github.com/kedgeproject/kedge/issues/555

### DIFF
--- a/scripts/update-json-schema.sh
+++ b/scripts/update-json-schema.sh
@@ -35,6 +35,6 @@ git config user.email "$GIT_EMAIL"
 git add "$OUTPUT_DIR"
 
 # Check if anything changed, and if it's the case, push to origin/master.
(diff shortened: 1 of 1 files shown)


scripts/update-json-schema.sh
```diff
@@ -34,7 +34,12 @@ git config user.name "$GIT_USER"
 git config user.email "$GIT_EMAIL"
 git add "$OUTPUT_DIR"
 
+# commit description
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
containscafeine: @cdrage $TRAVIS_PULL_REQUEST is always going to be false since this only work on merge to master. I think it's safe simply remove the PR number altogether to fix #555

cdrage: Yeah, that'd be better

cdrage:
Updated. @containscafeine

PR #577, opened 2018-01-19
Fix building docs

cdrage:
Fixes building the documentation since the "kedge" folder already
exists.


cdrage:
Docs aren't syncing. This will (try) to fix it.

PR #592, opened 2018-02-14
Added shortcut for route

surajnarwade:


surajnarwade:
@kadel @cdrage review needed :)

kadel:
test are failing :-(

surajnarwade:
@kadel tests are failing because `routeEndpoint` and `ingressEndpoint` is not present in JSON schema

kadel:
hmm :-(
this needs to be fixed either in this PR, or before this PR gets merged

surajnarwade:
@kadel , added tests

docs/file-reference.md
```diff
@@ -727,11 +727,12 @@ services:
...
-| Field        | Type            | Required | Description                       |
-|--------------|-----------------|----------|-----------------------------------|
-| name         | string          | yes      | The name of the service.          |
-| endpoint     | string          | no       | The endpoint of the service.      |
-| portMappings | array of "port" | no       | Array of ports. Ex. `80:8080/tcp` |
+| Field           | Type            | Required | Description                       |
+|-----------------|-----------------|----------|-----------------------------------|
+| name            | string          | yes      | The name of the service.          |
+| ingressEndpoint | string          | no       | The endpoint of the service.      |
+| routeEndpoint   | string or bool  | no       | The route endpoint of service
```
cdrage: missing a `|` at the end

pkg/spec/resources.go
```diff
@@ -417,7 +417,38 @@ func (app *App) createServices() ([]runtime.Object, error) {
...
+							Name: svc.Name,
+						},
+					},
+				}
+
+				switch port.RouteEndpoint {
+				case "true":
+					// OpenShift will take care of route
+					svcs = append(svcs, endpointRoute)
+				case "false":
```
cdrage: remove this case, don't need it if were just doing a debug

pkg/spec/resources.go
```diff
@@ -417,7 +417,38 @@ func (app *App) createServices() ([]runtime.Object, error) {
...
+					// Do not create route
+					log.Debug("Don't create route")
+				default:
+					// Route URL is given by user
+					endpointRoute.Spec.Host = port.RouteEndpoint
+					svcs = append(svcs, endpointRoute)
+
```
cdrage: i believe you can shorten this to an if statement instead.. ex:

```go
if port.RouteEndpoint {
 svcs = append(svcs, endpointRoute)
} else {
 endpointRoute.Spec.Host = port.RouteEndpoint
 svcs = append(svcs, endpointRoute)
}
```

cdrage (changes requested):

pkg/spec/resources.go
surajnarwade: @cdrage , since port.RouteEndpoint is string, I can't do that

docs/examples/route-endpoint/httpd-false.yml
```diff
@@ -0,0 +1,12 @@
...
+- containers:
+  - image: bitnami/nginx
+
+services:
+- name: httpd
+  type: NodePort
+  ports:
+    - port: 8080
+      targetPort: 8080
+      routeEndpoint: false
```
containscafeine: do we need an example with false?

docs/file-reference.md
```diff
@@ -778,6 +779,26 @@ of `service`.
...
+OR
+
+```yaml
+routeEndpoint: true
+```
+| Type           | Required | Description                  |
+|----------------|----------|------------------------------|
+| string or bool | no       | The endpoint of the service. |
+
+This is an added field in the Service port, which if specified an `route`
```
containscafeine: s/an/a

concaf (approved):

surajnarwade:
@containscafeine , @kadel can we merge this ?

Issue #601, opened 2018-02-26
Add `restartpolicy` and `imagepullpolicy` to `kedge init`


surajnarwade:
Add `--restart-policy` and `--image-pull-policy` args to `kedge init`

PR #603, opened 2018-02-27
Added restartpolicy and imagepullpolicy to kedge init

surajnarwade:
resolves issue #601

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -26,9 +26,10 @@ import (
 )
 
 var (
-	fileName, image, name string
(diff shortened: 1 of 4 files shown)


cdrage:
Is this good in your opinion to add to `kedge init`? It's a lot more flags to add when it's already simple.. @kadel

surajnarwade:
I am assuming, we will try to support most of the flags which are supported by `kubectl run`

surajnarwade:
cc @kadel @surajssd

PR #604, opened 2018-02-28
Improved help on error on kedge commands

surajnarwade:
For example,
```
kedge apply
No files were passed. Please pass file(s) using '-f' or '--files'
See 'kedge apply -h' for help and examples.
```

Similar to `kubectl`

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cmd
 
 import (
-	"fmt"
 	"os"
 
 	pkgcmd "github.com/kedgeproject/kedge/pkg/cmd"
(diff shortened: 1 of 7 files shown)


surajnarwade:
@containscafeine @cdrage @kadel does this make sense ?

cdrage:
@surajnarwade Why not use logrus?

surajnarwade:
@cdrage , I didn't get it, logrus where ?

surajnarwade:
@cdrage Kubernetes also uses same logic for this help

concaf (commented):
I like this approach. +1 from me once tests are fixed

